PR #11, opened 2015-10-25
Added classes Lambdify, LambdifyCSE, squashed from gh-3

bjodah:
This is a squashed commit for all of #3 (with minor edits) rebased on current master.

Things not addressed in this PR (which we possibly want to look into later):

- Testing with NumPy on appveyor
- Using py.test

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,7 +28,9 @@ build*
 *.pyc
 MANIFEST
 *.egg-info
-dist
+dist/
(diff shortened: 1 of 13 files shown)


certik:
Thanks @bjodah. I think this looks very good. Thanks for fixing all the issues!

certik:
I sent the minor fix in #15.

PR #13, opened 2015-10-26
return SymPy Rel when given symbolic relations; closes #12

### LINKED ISSUE #12: raise NotImplemented for symbolic relations
(heavily edited)
From https://github.com/symengine/symengine/issues/356

As long as relations are not implemented in symengine we should not return `True` or `False` for any symbolic inequality.

So at the moment the only way to conform to SymPy behaviour is by returning a SymPy Rel.

rwst:


rwst:
I don't get this. In IPython it's all fine...

certik:
The patch is fine with me to merge.

rwst:
Thanks!

PR #26, opened 2015-11-21
Add SymEngine ascii-art

Sumith1896:
Usage(as in Ruby wrappers):
```
>>> import symengine
>>> symengine.ascii_art()
```
Good thing to have in all our wrappers :)

### DIFF
--- a/symengine/__init__.py
+++ b/symengine/__init__.py
@@ -13,6 +13,15 @@ if have_mpc:
 
 __version__ = "0.1.0.dev"
 
+def ascii_art():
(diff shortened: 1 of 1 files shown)


certik:
In that case we should move it to the C++ library, so that if we change it (and I proposed to change it in some of our issues), we change it at one place.

Sumith1896:
Agreed, I'll try to do that.

Sumith1896:
Will update once this https://github.com/symengine/symengine/pull/797 is merged.

PR #35, opened 2016-01-04
Test https://github.com/symengine/symengine/pull/675

certik:


Sumith1896:
#675 merged.

certik:
We should update `symengine.py`, there are some unrelated failures which we should also fix.

PR #37, opened 2016-01-08
Avoid overloading diff

cbehan:
This fixes the build after a recent symengine change.


cbehan:
Oops, this is probably not a real fix. The wrapper I end up with cannot be imported.

Sumith1896:
@cbehan We should port the new code. Removing is not an option.

PR #39, opened 2016-01-27
Use multiset for derivatives

cbehan:
One of the handful of changes needed to get the bindings working again.


cbehan:
PR40 has a couple things that this was missing :).

certik:
I think a similar patch got in.

PR #41, opened 2016-01-28
Make Travis fail on error

certik:


certik:
We should implement erroring out for `install_travis.sh`, but this PR in particular is broken.

PR #53, opened 2016-02-15
Remove isolated print statement

cbehan:
It looks like this was left over from some debugging.


certik:
Indeed. +1 if tests pass.

PR #237, opened 2018-07-22
[WIP] Stepwise changes of gh-236 (to debug CI failiures)

bjodah:
This PR is for adding changes from gh-236 as multiple commits to find out where CI jobs start failing.


bjodah:
This is worrisome. Not sure what might be the cause of this.

PR #252, opened 2018-10-06
Update symengine hash and test gh-251

bjodah:


bjodah:
No new test failures (only the same old failure after `symengine.tests.test_sage.test_sage_conversions`).

PR #359, opened 2021-07-30
symengine_wrapper.pyx: Calculate the zero-th derivative

GSam:
```

This is useful for more general, albeit sometimes inefficient, summation
code. Just copying some taylor expansion code from the internet worked
with sympy but not symengine due to the following error:

  File "example.py", line 32, in <module>
    taylor(symengine.sympify('1/(1-x-x^2)'), 0, 15)
  File "example.py", line 27, in taylor
    p = p + (function.diff(x, i).subs(x, x0))/(factorial(i))*(x - x0)**i
  File "symengine_wrapper.pyx", line 923, in symengine.lib.symengine_wrapper.Basic.diff
  File "symengine_wrapper.pyx", line 4020, in symengine.lib.symengine_wrapper.diff
OverflowError: can't convert negative value to size_t

There are a number of inconsistencies compared to the regular sympy library:

1. Takes no argument:

   >>> symengine.sympify('x**2').diff()
   2*x

   This actually seems helpful, except when you have #2 misleading you.

   This is not accepted in sympy, but doesn't necessarily need a fix.

2. Takes a standalone integer argument and returns a nonsense answer:

   >>> symengine.sympify('x**2').diff(1)
   x**2

   This is simply not accepted in the regular sympy library.

3. Takes multiple integer arguments and aggregates:

    diff(x, 1, 1) == diff(x, 2)

   This does not occur with the sympy library, and does not seem useful.
   It can easily be replaced with diff(x, x), if repetition was necessary.

4. Takes only integer arguments > 1:

   diff(x, 0) returns integer overflow

   This is accepted in sympy.

In this patch, 2-4 are all fixed. #1 seems like a potentially useful (and
likely used) feature.
```


rikardn:
Thanks for submitting this. There are currently some general issues with the CI checks, which is probably why they fail in this PR.

The list of inconsistencies is also very useful. Perhaps some fixes could go into symengine directly.

PR #361, opened 2021-08-23
Update CI

rikardn:
* Stop testing using Python 3.6
* Test on Ubuntu 20.04
* Update to latest version of symengine


rikardn:
Thanks @isuruf. Much appreciated!

PR #370, opened 2021-09-24
[RFC] Add latex representation to DenseMatrix

eendebakpt:
- Add a `_repr_latex_` method to the `DenseMatrix` class. 
- Add truncation for large matrices

image

Related to https://github.com/symengine/symengine/issues/1701


certik:
Nice!

I think in general these things should go into the C++ symengine, and we should just call it from the Python wrappers. That way any language will benefit and we only have to maintain this at one place.

Let me know what you think. I can help you get started in the C++ codebase if you are interested.

eendebakpt:
> Nice!
> 
> I think in general these things should go into the C++ symengine, and we should just call it from the Python wrappers. That way any language will benefit and we only have to maintain this at one place.
> 
> Let me know what you think. I can help you get started in the C++ codebase if you are interested.

If there would be other languages that benefit from latex it might indeed be good to move this into the `symengine` part. Otherwise it is probably easier to implement it on the python side.
I am comfortable with C++, but some pointers into the `symengine` structure would be helpful. In particular: I think the basic elements of a `DenseMatrix` are `Basic` objects. Where is the (latex) printing for the `DenseMatrix` and `Basic` objects? And which other structures to consider.

The latex printing of simple objects from the python side also seems not to be implemented, although a `_repr_latex_` is present:
image

isuruf:
> The latex printing of simple objects from the python side also seems not to be implemented, although a _repr_latex_ is present:

You need to do `init_printing()` to make it work,

eendebakpt:
> > The latex printing of simple objects from the python side also seems not to be implemented, although a _repr_latex_ is present:
> 
> You need to do `init_printing()` to make it work,

Thanks for the pointer. On my system it is not working for some reason:
image
I will try a fresh install in a new env.

certik:
Try this in a jupyter notebook. The `init_printing` should work.

certik:
@isuruf, since DenseMatrix is not part of the SymEngine visitor pattern, and currently does not have a latex printer, what do you think is the best design for printing latex? In `symengine/printers.h` we have `std::string latex(const Basic &x);` but that accepts `Basic`. I wonder if we can add another overload like `latex(const DenseMatrix &x);`, add an implementation into `symengine/printers/latex.cpp` and then just call it from the Python wrappers.

isuruf:
> In symengine/printers.h we have std::string latex(const Basic &x); but that accepts Basic. I wonder if we can add another overload like latex(const DenseMatrix &x);, add an implementation into symengine/printers/latex.cpp and then just call it from the Python wrappers.

Yes, that sounds good.

eendebakpt:
@isuruf I updated this PR based on https://github.com/symengine/symengine/pull/1847. Could you help me out with the `cython` part where the python object is mapped to a c++ symengine object?

peendebak:
@isuruf The CI is failing because a python object cannot be assigned to the ctypes `DenseMatrix`. Could you have a look at it?

isuruf:
Thanks

PR #423, opened 2022-10-09
Use print() function in both Python 2 and 3

cclauss:
Old-style `print` expressions are a **syntax errors** in Python3

`print()` function works as expected on both Python 2 and Python 3.

### DIFF
--- a/benchmarks/expand2_sage.py
+++ b/benchmarks/expand2_sage.py
@@ -1,9 +1,10 @@
+from __future__ import print_function
 from timeit import default_timer as clock
 from sage.all import var
 var("x y z w")
 e = (x+y+z+w)**15
 f = e*(e+w)
-print f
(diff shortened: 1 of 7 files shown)


rikardn:
Good catch! I don't think symengine.py is supporting Python 2 any longer so everything should be Python 3

cclauss:
Yes.  Python 2 died more than 1,000 days ago on 1/1/2020 so there are almost no active projects that still support it.

rikardn:
Then we wouldn't need the `from future` import

cclauss:
That can be automated with [`pyupgrade --py37-plus **/*.py`](https://github.com/asottile/pyupgrade#__future__-import-removal)

rikardn:
Thanks for the tip and for the PR.

PR #427, opened 2022-11-10
ci: Add Python 3.11 to the testing

cclauss:
Related to #422, #425


rikardn:
Thanks for the PR!

cclauss:
Does the Python 3.11 performance boost effect the benchmarks or do they a measure of C++ performance?

PR #446, opened 2023-06-24
Tentative fix of .has for FunctionSymbol

bjodah:
Tests pass locally.

During debugging I tried to `from symengine import FunctionSymbol`, I had too look into `__init__.py` to find it was renamed to `AppliedUndef`, so I also changed `__init__.py` to simply add this as an alias instead. Also enabled a newer CMake policy (allows to set SymEngine_ROOT as an environment variable).


isuruf (approved):

bjodah:
Thank you for the review @isuruf

PR #454, opened 2023-11-05
GitHub Action to pip install symengine on current Python and PyPy

cclauss:
`pip install symengine` ___fails___ on Python 3.12 and PyPy 3.10
* https://github.com/symengine/symengine.py/actions/workflows/pip_install.yml
* #421
---
https://github.com/symengine/symengine.py/blob/810ef475f34388bb63ae366bf6b238762b5c2d62/setup.py#L14
https://packaging.python.org/en/latest/current.html#packaging-tool-recommendations raises a 404 Not Found but https://packaging.python.org works.

### DIFF
--- a/.github/workflows/pip_install.yml
+++ b/.github/workflows/pip_install.yml
@@ -10,6 +10,7 @@ jobs:
         python-version: ['3.12', 'pypy3.10']  # Other versions above install successfully
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


cclauss:
```
Could not find a package configuration file provided by "SymEngine"
        (requested version 0.8.1) with any of the following names:
      
          SymEngineConfig.cmake
          symengine-config.cmake
```

PR #478, opened 2024-05-21
build(cmake): avoid the usage of distutils

Moraxyc:
distutils has been removed since python 3.12


bjodah:
Thanks! (I notice that our CI-scripts needs some attention, I get the same failures in an unrelated PR).

PR #482, opened 2024-05-28
Bump one CI config to ubuntu-24.04

bjodah:
ubuntu-20.04 no longer available on github actions


bjodah:
I see, I missed that part, thank you for pointing that out.

Issue #474, opened 2024-02-21
PDM and Poetry can't install symengine properly


martinruefenacht:
I am trying to run symengine in a PDM or Poetry project and while I can install it via "pdm add" or "poetry add" and I find it in the virtualenv site-packages, the lib directory is empty compared to my globally installed version with my OS python.

When I run python directly from the venv and import symengine I get:

```
Python 3.12.1 (main, Dec 18 2023, 00:00:00) [GCC 13.2.1 20231205 (Red Hat 13.2.1-6)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import symengine
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/****/****/****/.venv/lib/python3.12/site-packages/symengine/__init__.py", line 12, in <module>
    import symengine.lib.symengine_wrapper as wrapper
ModuleNotFoundError: No module named 'symengine.lib.symengine_wrapper'
```

Is this a problem with both Poetry and PDM or with symengine not picking something up?

martinruefenacht:
Yes, I got the same with just a normal venv.

What I notice is that several things in the site-packages was missing in the pdm version:

```
symengine
├── functions.py
├── __init__.py
├── lib
│   ├── __init__.py
│   └── __pycache__
│       └── __init__.cpython-312.pyc
├── printing.py
├── __pycache__
│   └── __init__.cpython-312.pyc
├── sympy_compat.py
├── tests
│   ├── __init__.py
│   ├── test_arit.py
│   ├── test_cse.py
│   ├── test_dict_basic.py
│   ├── test_eval.py
│   ├── test_expr.py
│   ├── test_functions.py
│   ├── test_lambdify.py
│   ├── test_logic.py
│   ├── test_matrices.py
│   ├── test_ntheory.py
│   ├── test_number.py
│   ├── test_pickling.py
│   ├── test_printing.py
│   ├── test_sage.py
│   ├── test_series_expansion.py
│   ├── test_sets.py
│   ├── test_solve.py
│   ├── test_subs.py
│   ├── test_symbol.py
│   ├── test_sympify.py
│   ├── test_sympy_compat.py
│   ├── test_sympy_conv.py
│   └── test_var.py
├── test_utilities.py
└── utilities.py
```

The lib/ folder is empty, symengine.libs doesn't get created, and symengine-0.11.

Perhaps this is more a question for pdm and poetry than symengine? They seem to be missing the external libraries.

[...]

PR #510, opened 2025-03-31
Fix Fedora/RedHat installation location (#474)

aostrowski-hbn:
Those OSes have separate platlib and purelib directories. Symengine was installed partially into both from version 0.10.0 onwards. 

This change marks some of it's packages as ext_modules, which makes setuptools treat the wheel as platform specific and install if fully into platlib (lib64) instead of purelib (lib), which should solve #474.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 2.8.12...4.0.0)
 
 if (POLICY CMP0057)
   cmake_policy(SET CMP0057 NEW) # needed for llvm >= 16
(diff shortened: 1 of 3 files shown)


aostrowski-hbn:
> Thanks for the fixes. The CI fails with:
> 
> ```
> + python3 /home/runner/work/symengine.py/symengine.py/bin/test_python.py
> Traceback (most recent call last):
>   File "/home/runner/work/symengine.py/symengine.py/bin/test_python.py", line 4, in <module>
>     import symengine
>   File "/home/runner/our_usr/lib/python3.10/site-packages/symengine/__init__.py", line 12, in <module>
>     import symengine.lib.symengine_wrapper as wrapper
> ImportError: dynamic module does not define module export function (PyInit_lib)
> ```
> 
> Is this error caused by this PR, or is it in master already?

It was just in my PR. I see the change setting `SYMENGINE_INSTALL_PY_FILES=ON` solves the issue.

PR #525, opened 2025-11-24
More explicit testing of pickling of user defined symbol

bjodah:
This draft-PR tracks symengine/symengine#2129


bjodah:
So when using:
https://github.com/symengine/symengine/pull/2129/commits/d10ee0be16d25f1e35d5cd31d6ac2788eb426833

we get this pytest failure consistently:
```python
________________________________ test_pysymbol _________________________________

    def test_pysymbol():
        a = MySymbol("hello", attr=1)
        b = pickle.loads(pickle.dumps(a + 2)) - 2
        c = MySymbol("hello", attr=-1)
        try:
>           assert a == b
E           assert hello == hello

symengine/tests/test_pickling.py:43: AssertionError
```
(b is no longer a `MySymbol` instance but simply an instance of `Symbol`). As to why this happens: I'm currently clueless, my understanding was that the information regarding the python instance would be by the pickling mechanism (i.e. *outside* of the serialized c++ representation), but clearly I'm missing something.

bjodah:
Nevermind, here's the reason:
https://github.com/symengine/symengine.py/blob/4f0775aa14707586b6b8bdac1f3946ed2c833cda/symengine/lib/pywrapper.cpp#L327

PR #527, opened 2026-06-14
Support set_basic and multiset_basic with _GLIBCXX_DEBUG

bjodah:
So Cython's [declaration](https://github.com/cython/cython/blob/765d9b62b439ceae4b132339dfec18f52d7adc07/Cython/Includes/libcpp/set.pxd#L4) of `std::set` only defines a single template argument, which breaks `g++ -D_GLIBCXX_DEBUG`.

A secondary benefit of this patch is a looser coupling with exact underlying datatype used in SymEngine (at least it looks so to me).

I've been applying this patch locally for a while. If you want, I could add modify/append a config in the CI build matrix to actually exercise g++ with debug enabled standard library.


isuruf:
Thanks

